PR #6, opened 2015-01-07
Remove extra "e" in include directive.

jhunold:
Some typo or mistake introduced while resolving merge conclicts.maybe. This breaks windows compilation only.


olk:
thx

PR #11, opened 2015-04-16
Changes required for aarch64 support in boost::context.

ajasty-cavium:
Request review of changes required for boost::context support on aarch64.


olk:
- detail/fcontext_arm.hpp (etc.) are not used anymore (fcotnext_t is obaque pointer)
- ARM uses the AAPCS calling convention not SYSV (boost.build should already set os.platform to ARM, see https://github.com/boostorg/build/pull/58)

PR #12, opened 2015-04-23
have jump,make_fcontext show up as importable

haubi:
On AIX, building with aix-soname=svr4, nm is used to list public symbols
in object files. This fixes the assembly to have make_fcontext and
jump_fcontext show up as public symbols.

The problem actually shows up when linking the coroutine library against the context library when built with aix-soname=svr4. 
However, except for this patch allowing to compile and link even coroutine, I'm unsure if the assembly is correct from all aspects, like alignment, TOC, and whether it works.

### DIFF
--- a/src/asm/jump_ppc32_sysv_xcoff_gas.S
+++ b/src/asm/jump_ppc32_sysv_xcoff_gas.S
@@ -1,5 +1,9 @@
 .globl .jump_fcontext
+.globl  jump_fcontext[DS]
 .align 2 
+.csect	jump_fcontext[DS]
+jump_fcontext:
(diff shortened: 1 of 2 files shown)


olk:
thx, I've applied the patch to branch develop

xantares:
hi @haubi, @olk,
I'm experiencing a similar issue with boost 1.58 for mingw64:
an error showing undefined refs to make_fcontext & jump_fcontext whe building a cpp from coroutine
I use this patch: http://pkgs.fedoraproject.org/cgit/mingw-boost.git/tree/boost-mingw.patch
hwo can I fix it for mingw too ?

PR #13, opened 2015-04-25
Do not export symbols for static build on PE builds.

dlunch:
Symbols incorrectly marked for export on static libraries exported on final executable. This change make symbols not to be marked as export on static builds.


olk:
thx

PR #18, opened 2015-09-01
Fix several typos I spotted while reading the documentation

BenWiederhake:
Nothing major, just typos. I used `grep` and manual replacement to fix duplicate typos. These typos were spotted while reading the Boost.Coroutine doc; see this PR: https://github.com/boostorg/coroutine/pull/22

I don't think that I'll spot more typos soon. If I do, I'll behave like the Nagle's algorithm and will *not* open another pull request.


olk:
thx

PR #19, opened 2015-09-17
Fix signed pointer comparison.

BenWiederhake:
I created this using find, grep, and sed. No manual editing, but lots of manual inspection. I'm pretty sure that this is correct, and I'm currently running all regression tests. I'll follow up on whether they are successful or not, but just wanted to provide grounds for discussion for https://github.com/boostorg/coroutine/issues/23 .


olk:
thx

BenWiederhake:
After some initial problems with generally building it at all (libbz-dev was missing), all regression tests pass on my machine.

PR #25, opened 2016-02-05
Don't link boost.thread if <mutex> is available.

Flast:
Linking boost.thread is unnecessary if the std library supports std::call_once.

Tested: Fedora Rawhide x86_64 / GCC 6.0.0 20160201


olk:
nice, thx

PR #26, opened 2016-03-24
Fix boost::once_flag initialization

Lastique:
The initial value is required to be BOOST_ONCE_INIT. Luckily, it is currently equivalent to zero, that's why it worked before.


olk:
thx

PR #27, opened 2016-03-25
Don't use Boost.Config build-time configure target.

Flast:
In some cases it doesn't respect config macros.

I believe this PR will fix build problem reported on devel ML but I can't test because I have no Mac (or similar environment).

Any comments and test results are welcome.


olk:
thank you for fixing this issue

PR #34, opened 2016-10-02
Fix compilation in c++1z

blastrock:
invoke() could resolve to both boost::context::detail::invoke and std::invoke
because of ADL, which would result in an ambiguous call error

This patch was tested on debian unstable, with gcc 6.2 and clang 3.9. Also, I didn't test the patch in this repository but directly on boost 1.61, but it should be safe enough.


olk:
thx - I'll add defect macro for boost.config regrading to std::invoke() soon

Issue #50, opened 2017-04-05
Bad assembly for fcontext on ppc64/sysv/elf


DaoWen:
This past week I was testing some code (which uses Boost.Context routines) on a POWER7 RHEL system. I had hundreds of successful test runs on x86_64 machines, but I was getting segfaults everywhere on the ppc64 system. I traced the problems back to the machine code for the `fcontext` functions, where I identified several bugs in the assembly routines:

  1. Saving `r29` into `r28`'s spot (typo).
  2. Assuming identical ABI for returning `transfer_t` struct and passing the struct as the first argument to a function.
  3. No distinction between ELF v1 and v2 in the `ontop_fcontext` when jumping into the target function.

These bugs were introduced in v1.61, which is when the internal `fcontext` interface was updated. I didn't have any problems previously when using v1.54. I've patched the `ppc64_sysv_elf_gas` versions of the `fcontext` functions, and I'll open a pull request shortly; however, the same bugs are also present in the code for other ppc targets.

I was surprised to find so many bugs in the code. When I ran the unit tests on the POWER7 machine, they failed consistently. This leads me to think this code was not actually tested. I understand that the library maintainers might not have the resources to test the code on all of the "supported" platforms; however, maintaining some sort of a testing matrix in the documentation listing which versions have been verified would be useful. Adding an obvious warning to check the verification status for platforms with lower-priority support would also be helpful.

olk:
I've no access to to POWER systems - so I can't test the code. So I've to wait for bug reports.
Unfortunately sometimes I get false success replies from users.

ty for the fix

olk:
Coukld you fetch the latest sources and run the unit-tests for PPC32 + PPC64 please.

DaoWen:
Yes, I should be able to make some time to do that tomorrow.

DaoWen:
I tested the latest sources with ppc64/sysv/elfv1, and as far as I can tell all of the tests are passing. Unfortunately, I don't have the dependencies installed to build 32-bit applications with GCC. I'll see if I can get that set up later and then test the 32-bit version.

olk:
thank you - that's really nice

PR #51, opened 2017-04-05
Fix assembly for fcontext on ppc64/sysv/elf

DaoWen:
Fixes bugs in the `fcontext` assembly functions for the ppc64/sysv/elf platform.

See issue #50 for more details.

This patch was tested against the latest `develop` commit (a429ff6301) using GCC 5.4 on a POWER7 machine running RHEL 6.6. However, I am not at all familiar with the Boost built/test system, so I am not entirely confident that I ran the unit tests correctly. I ran the following command from the Boost project root directory:

    ./b2 cxxflags="-std=c++11" --build-dir="/tmp/nick/boost/build" --with-context libs/context/test

The tests ran and failed on the first test without the patch, and then several tests ran and all passed after applying the patch. I think that means the patch works correctly (as far as the functionality covered by those tests goes anyway). Note that since my platform only uses ELFv1, I was not able to verify the ELFv2-specific code.

### DIFF
--- a/test/test_fcontext.cpp
+++ b/test/test_fcontext.cpp
@@ -9,6 +9,7 @@
 #include <stdexcept>
 #include <string>
 #include <utility>
+#include <cstdio>
 
 #include <boost/array.hpp>
 #include <boost/assert.hpp>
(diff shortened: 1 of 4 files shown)


test/test_fcontext.cpp
```diff
@@ -9,6 +9,7 @@
 #include <stdexcept>
 #include <string>
 #include <utility>
+#include <cstdio>
```
DaoWen: I get an error without this include: *`sscanf` is not a member of namespace `std`*

olk:
ty

PR #53, opened 2017-04-18
Fix register usage for ppc64/sysv/elfv2

DaoWen:
I was able to set up some PowerPC virtual machines using QEMU, and I tested the previous patch on ppc (32-bit), ppc64 big-endian (ELFv1), and ppc64 little-endian (ELFv2). The previous patch didn't work with ELFv2 (significant ABI differences from ELFv1). This version was tested and passes the unit tests on all three of the virtual machines mentioned above, although I was only able to run the fcontext test on the 32-bit ppc target since I was using an older version of GCC.


olk:
that's very nice, ty

PR #55, opened 2017-04-20
Fix callcc functionality for g++-4.9

DaoWen:
Fixes the resolution issue discussed in #52 and #54 that caused compiler errors when using g++-4.9. Also fixes a bad cast that was causing bad values to be passed as arguments into continuations.

### DIFF
--- a/include/boost/context/continuation.hpp
+++ b/include/boost/context/continuation.hpp
@@ -56,7 +56,7 @@ namespace boost {
 namespace context {
 namespace detail {
 
-template< int N >
+template<typename U>
 struct helper {
     template< typename T >
     static T convert( T && t) noexcept {
@@ -64,8 +64,8 @@ struct helper {
     }
 };
 
-template<>
-struct helper< 1 > {
+template<typename U>
+struct helper< std::tuple<U> > {
     template< typename T >
     static std::tuple< T > convert( T && t) noexcept {
         return std::make_tuple( std::forward< T >( t) );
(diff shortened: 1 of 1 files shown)


include/boost/context/continuation.hpp
```diff
@@ -451,7 +452,8 @@ class continuation {
 template<
     typename Fn,
     typename ... Arg,
-    typename = detail::disable_overload< continuation, Fn >
+    typename = detail::disable_overload< continuation, Fn >,
+    typename = detail::disable_overload< std::allocator_arg_t, Fn >
```
DaoWen: This extra `disable_overload` fixes the compiler error for me.

include/boost/context/continuation.hpp
```diff
@@ -354,15 +355,15 @@ class continuation {
     template< typename Fn, typename ... Arg >
     continuation resume_with( Fn && fn, Arg ... arg) {
         BOOST_ASSERT( nullptr != t_.fctx);
-        auto tpl = std::make_tuple( std::forward< Fn >( fn), std::forward< Arg >( arg) ... );
```
DaoWen: Note that the type of `tpl` here is `std::tuple(Fn, Arg...)`.

include/boost/context/continuation.hpp
```diff
@@ -239,15 +239,16 @@ struct result_type< Arg > {
 
 }
 
-template< typename Ctx, typename Fn, typename ... Arg >
+template< typename Ctx, typename Fn, typename Arg >
 detail::transfer_t context_ontop( detail::transfer_t t) {
-    auto p = static_cast< std::tuple< Fn, std::tuple< Arg ... > > * >( t.data);
```
DaoWen: Note that the expected type for `t.data` is `std::tuple< Fn, std::tuple< Arg ... > > *`, which does not correspond to the argument type declared on line 357 below.

My guess is that the C++ library implementation for `std::tuple` changed between g++-4.9 and g++-5, which is why this bad cast happened to work with newer g++ versions but fails with version 4.9.

olk:
hmm, did noticed that mistake - strage that I got fooled by so much compilers (clang, msvc, intel compile this code).
ty for fixing it.

Issue #62, opened 2017-07-28
jump_fcontext crashes in 32bit iOS simulator


mjohns:
jump_fcontext crashes in jump_i386_sysv_macho_gas.S when trying mov the "to" fctx parameter into eax. This is easy to reproduce on a mac by compiling a simple program with the assembly files and a simple main function. (This example works with the old assembly files before the refactoring to have transfer_t etc)

With a little debugging it seems like the offsets to get the "to" and "data" parameters are not correct anymore. The seem to be off by 4 bytes.

Compile with:  g++ -g *.S main.cc -m32 -std=c++11

#include <cstdio>
#include <cstdlib>

typedef void* fcontext_t;

struct transfer_t {
  fcontext_t fctx;
  void* data;
};

extern "C"
transfer_t jump_fcontext(fcontext_t const to, void* vp);
extern "C"
fcontext_t make_fcontext(void* stack_pointer, size_t size,
                                           void (*fn)(transfer_t));

void Run(transfer_t t) {
  printf("not reached");
}

int main() {
  const size_t size = 131072;
  char* smem = (char*)malloc(size);
  void* sp = smem + size;
  
  fcontext_t fctx = make_fcontext(sp, size, &Run);
  // jump_fcontext crashes.
  jump_fcontext(fctx, nullptr);
}

olk:
boost's regression (develop - http://www.boost.org/development/tests/develop/developer/context.html) tests for context/coroutine/coroutine2/fiber show no errors on MacOS X.
Note that jump_fcontext is a internal API and should not be used by users (not supported).

seams to me this bug report is invalid

mjohns:
Are there actually 32bit MacOS tests? I only see one test for darwin where windows has multiple 32bit ones listed. It is admittedly a rare platform since all osx machines are typically 64bit

mjohns:
I was able to build 32bit osx boost and use the real execution_context interface to reproduce the same crash. I don't think the i386_macho_gas assembly files are exercised in the regression tests.

Here is the example program that crashed in the same way (taken from one of the regression tests):
`
#include <boost/context/execution_context_v2.hpp>
#include <cstdio>

namespace ctx = boost::context;

double value3 = 0.0;

ctx::execution_context< void > fn3( double d, ctx::execution_context< void > && ctx) {
  d += 3.45;
  value3 = d;
  return std::move( ctx);
}

int main() {
  double d = 7.13;
  ctx::execution_context< void > ctx( fn3, d);
  ctx();

[...]

PR #64, opened 2017-08-02
Fix i386/macho routines to correctly return transfer_t in EAX/EDX

mjohns:
The OSX 32bit ABI differs from the System V ABI in that 8 byte structs
are returned in registers EAX/EDX instead of passing the address of the
struct on the stack. See: https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/LowLevelABI/130-IA-32_Function_Calling_Conventions/IA32.html

The current implementation for this platform crashes when compiled for
32bit OSX (e.g. when running a 32bit iOS simulator) due to not
satisfying the ABI.


mjohns:
issue #62

olk:
very nice, ty

PR #68, opened 2017-09-22
use perfect forwarding for StackAlloc

z33ky:
In a project of mine I have a non-copyable StackAlloc-type. I changed StackAlloc to use perfect forwarding so that it can be moved instead of copied.


olk:
fine, ty

PR #77, opened 2018-03-16
Added working compiler versions and features needed for the lib

teeks99:
Better details on what the compiler requirements are for building/using context.


olk:
ty

PR #85, opened 2018-09-14
Duplicate alias should be missing GCC alias

kuhlenough:
Turn duplicate sysv-power-32-elf-clang alias into missing sysv-power-32-elf-gcc alias


olk:
ty

PR #90, opened 2018-11-09
Build/remove useless lines

### LINKED ISSUE #76: Missing assembly files on iOS
Hello,

I had some trouble building this library on iOS (armv7/armv7s/x86/x86_64/armv8) and had to patch the build system files in order to add the relevant missing files.

I found this [SO thread](https://stackoverflow.com/a/32501078) that saved me days of research, and I came up with the following patches: 
[patches.zip](https://github.com/boostorg/context/files/1803121/patches.zip)

Having no notion of Boost.Build, I monkey-patched the file once per arch, and have to apply the correct patch each time (not doing it by hand though).

I was thinking that you would have a better fix than me, and having it fixed upstream would be great!

theodelrieu:
After some research, it seems that there is nothing wrong with the `Jamfile.v2`, I just got confused by some unused code. Thus I'm removing them to avoid confusion in the future.

Fixes #76


olk:
ty

PR #95, opened 2018-12-16
Fix bad static_cast in resume_with:

djarek:
- fix a cast from `void*` to the wrong dynamic type. The code cast from
  `tuple<decay_t<Fn>>*` -> `void*` -> `tuple<Fn>*`, which worked when an rvalue
  was passed to resume_with(), but broke if the function was passed
  by lvalue-reference, because it resulted in a cast from `Fn` to `Fn&`.
- remove use of `std::tuple` - no need to instantiate it for every callable.


olk:
ty

PR #119, opened 2019-09-20
Optimize s390x instructions

nealef:
Some optimizations for the s390x implementation of jump/make/ontop:
- Use multiple register instructions rather than individual load/stores
- Use floating point instructions to load or store from/to storage
- Use symbols to map CTX rather than explicit displacements
- Branch to _exit@PLT to avoid relocation complaints
- Use immediate instructions for adding to registers

Builds and passes fcontext tests using gcc-4.8.5 and devtoolset-8-gcc-c++ (8.3.1).

Do you need access to a jenkins worker so you can build/test s390x automatically?

### DIFF
--- a/src/asm/jump_s390x_sysv_elf_gas.S
+++ b/src/asm/jump_s390x_sysv_elf_gas.S
@@ -36,47 +36,34 @@
 .global jump_fcontext
 .type   jump_fcontext, @function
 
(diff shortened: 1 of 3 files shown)


olk:
very nice - ty
boost uses Travis CI - unfortunately the code is not build and executed on a s390x system

PR #127, opened 2019-11-22
Fix exports for 32-bit GNU assembler files targeting Windows.

ericastor:
The incorrect exports can cause linker issues (at least when using LLVM's lld-link) claiming that these primitives are not available.


olk:
ty

PR #128, opened 2019-11-28
Mark 32-bit GNU assembly files targeting Windows as safeseh compatible

ericastor:
We set the LSB of the magic symbol @feat.00 to 1. This symbol is used for compiler-linker communication, and the LSB expresses that an object file has opted into "safeseh"; any SEH handlers used in this file must be listed in the .sxdata section.

Since we don't register any SEH handlers in these files, this is trivially satisfied.

Reference: [the PE-COFF specification](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#the-sxdata-section).


olk:
ty

PR #129, opened 2019-12-06
Fix non-PIC in RISC-V assembly

andreas-schwab:
Since this ends up in a shared library we should use a PLT call.  Also,
the finish label is local, so we can use lla instead of la.


olk:
ty

PR #131, opened 2019-12-18
Use OSPLAT MIPS32/MIPS64 to set different ABI

wzssyqa:
We should set abi for 64bit as n64, and 32bit for o32.
This patch depends on the patch of boost/build.


wzssyqa:
This PR depends on: https://github.com/boostorg/build/pull/517

olk:
Ty

PR #132, opened 2019-12-18
mips64/n64: .align 3

wzssyqa:
The right align for mips64/n64 is 3 instead of 2.


olk:
Does it pass the unit tests?

wzssyqa:
> 
> 
> Does it pass the unit tests?

Yes. set .aligen here seem have no effect at all.
Nothing happens after the change value: the unit test have the same result as before.
Anyway, 3 is the right value.

FlyGoat:
It was my mistake, although it didn't cause any problem.
Variable should be aligned to 64bit boundary here in 64bit architecture.
Thanks for fixing.

olk:
Ty

PR #135, opened 2020-04-03
Fix gcc -Wunused-variable argument when compiling with -DNDEBUG (and …

Romain-Geissler-1A:
…without -DBOOST_DISABLE_ASSERT).


olk:
ty

PR #140, opened 2020-04-25
Changes for Embarcadero C++ clang-based compilers, targeting Boost 1.74

eldiener:


eldiener:
The Travis failure says:

"Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 clang-5.0 : Depends: libjsoncpp0 (>= 0.6.0~rc2) but it is not installable
E: Unable to correct problems, you have held broken packages."

This has nothing to do with this PR.

olk:
It is a problem of the build environment.

olk:
Is your pull request complete (assembler and configuration in the Jamfile) and do the unit-tests pass?

eldiener:
The pull request is complete.

olk:
But did you run the unit-tests and they succeed?

eldiener:
I do not know what you mean by "run the unit tests". Is not that what your CI tests are all about ? Do you mean did I test the Embarcadero C++ clang-based compilers ? I was able to build Boost context  with Embarcadero C++ clang-based compilers.

olk:
In sub directory _test_ execute b2. This runs the tests that CI executes too.

eldiener:
The Embarcadero clang-based compilers, which are currewntly based off of clang-5.0, have BOOST_NO_CXX11_THREAD_LOCAL defined since the clang--5.0 compiler, and therefore the Embarcadero compiler, put out an error whenever a thread local variable is a non-trivial class.

olk:
Why not test for BOOST_NO_CXX11_THREAD_LOCAL and use __thread etc.?

eldiener:
I know that the config test for thread local for Embarcadero clang-based compilers fails. I am not sure want you want me to do vis-a-vis context.

olk:
I suggest you wrap `thread_local` by testing for BOOST_NO_CXX11_THREAD_LOCAL.
If BOOST_NO_CXX11_THREAD_LOCAL is not defined use `thread_local` otherwise use the compiler specific TLS specifier (for GCC it's `__thread`).

eldiener:
BOOST_NO_CXX11_THREAD_LOCAL is defined for the Embarcadero clang-based compilers because it fails the thread local test in Boost.Config. I have reported this as a bug to Embarcadero ( https://quality.embarcadero.com/browse/RSP-28277?filter=-2 ). There is no compiler specific solution to this for the Embarcadero clang-based compilers until they fix their bug. Nonetheless I think it would be good to merge this PR so when they fix the bug context should work for the Embarcadero clang-based compilers. This PR does not affect any other compiler other than the Embarcadero clang-based compilers.

PR #145, opened 2020-06-28
Fixes for x32 ABI.

hvdijk:
In the x32 ABI, pointers take up 4 bytes, so a structure containing two pointers is passed and returned in a single 8-byte register.

Changes required to build for the x32 ABI in the first place are not part of this pull request: that is the subject of https://github.com/boostorg/build/issues/388. I have tested this PR on top of local modifications that bypass Boost's forced overriding of compiler flags.


olk:
ty

Issue #136, opened 2020-04-16
[1.73.0] i686-w64-mingw32/bin/ld: cannot export _jump_fcontext: symbol not found


jschueller:
the 1.73 version fails to build on the 32 bits variant (but succeeds on 64bits) of mingw with the following error:
```
gcc.link.dll bin.v2/libs/context/build/gcc-mingw64/release/binary-format-pe/target-os-windows/threading-multi/visibility-hidden/libboost_context.dll.a

    "i686-w64-mingw32-g++"   "-Wl,--out-implib,bin.v2/libs/context/build/gcc-mingw64/release/binary-format-pe/target-os-windows/threading-multi/visibility-hidden/libboost_context.dll.a" -o "bin.v2/libs/context/build/gcc-mingw64/release/binary-format-pe/target-os-windows/threading-multi/visibility-hidden/libboost_context.dll" -Wl,-h -Wl,libboost_context.dll -shared -Wl,--start-group "bin.v2/libs/context/build/gcc-mingw64/release/binary-format-pe/target-os-windows/threading-multi/visibility-hidden/asm/make_i386_ms_pe_gas.o" "bin.v2/libs/context/build/gcc-mingw64/release/binary-format-pe/target-os-windows/threading-multi/visibility-hidden/asm/jump_i386_ms_pe_gas.o" "bin.v2/libs/context/build/gcc-mingw64/release/binary-format-pe/target-os-windows/threading-multi/visibility-hidden/asm/ontop_i386_ms_pe_gas.o" "bin.v2/libs/context/build/gcc-mingw64/release/binary-format-pe/target-os-windows/threading-multi/visibility-hidden/dummy.o" "bin.v2/libs/context/build/gcc-mingw64/release/binary-format-pe/target-os-windows/threading-multi/visibility-hidden/windows/stack_traits.o"  -Wl,-Bstatic  -Wl,-Bdynamic  -Wl,--end-group -m32 -mthreads -fvisibility=hidden -fvisibility-inlines-hidden -march=i686 -Wl,-O1,--sort-common,--as-needed -fstack-protector

/usr/lib/gcc/i686-w64-mingw32/9.3.0/../../../../i686-w64-mingw32/bin/ld: cannot export _jump_fcontext: symbol not found
```

The faulty commit is https://github.com/boostorg/context/commit/85783e8cbba03804e13b3d314112df6b089ed2c0 from #127
Maybe fixing for LLVM's ld resulted in breaking the build with GNU's ld.

[...]

PR #150, opened 2020-08-24
Add variants for clang targeting Windows without MASM support

ericastor:
This is not an officially-supported Boost toolset, but is consumed by at least one user.

See: https://github.com/boostorg/context/issues/136#issuecomment-677695699


olk:
ty

PR #151, opened 2020-09-08
Use integer math instead of floating-point to compute number of pages

lowsfer:
The previous math can go wrong if size_ cannot be exactly expressed with fp32.

### DIFF
--- a/include/boost/context/posix/protected_fixedsize_stack.hpp
+++ b/include/boost/context/posix/protected_fixedsize_stack.hpp
@@ -51,10 +51,7 @@ public:
 
     stack_context allocate() {
         // calculate how many pages are required
-        const std::size_t pages(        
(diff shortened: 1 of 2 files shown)


lowsfer:
The error message for the failed test is not caused by this change. Looks like an infra issue:

> The following packages have unmet dependencies:
> 
>  clang-5.0 : Depends: libjsoncpp0 (>= 0.6.0~rc2) but it is not installable
> 
> E: Unable to correct problems, you have held broken packages.
> apt-get.diagnostics
> 
> apt-get install failed
> 
> The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install clang-5.0 libstdc++-7-dev" failed and exited with 100 during .

olk:
ty

PR #155, opened 2020-12-16
In cygwin we want to use UCONTEXT

viric:
Crash otherwise.

I haven't tried winfibers.

I think this should fix https://github.com/boostorg/coroutine2/issues/34. Is this the kind of patch that could get in? Or how else?


olk:
It is a workaround - better boost.context would be patched with correct assembler.

olk:
ty

PR #156, opened 2020-12-17
TSAN annotations

KochetovNicolai:
Annotate `ucontext` implementation using TSAN fiber switching API.


KochetovNicolai:
I have used `BOOST_USE_TSAN` macro for now. Don't know what is the proper way to do it.

olk:
ty

PR #159, opened 2021-02-13
fix softfloat support for MIPS64 ASM

neheb:
Added define also found in the MIPS32 ASM files.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


olk:
ty

PR #160, opened 2021-03-01
Add arm and arm64 into combined asm codes

dixyes:
Apple silicon M1 uses arm64 architecture, this commit will add arm and arm64 architecture into combined asm code in order to support newer apple's machine in combined asm codes.


olk:
ty

PR #168, opened 2021-05-05
Fix missing semicolon

ctrysbita:
The error was introduced in commit 3f3d358 and present in boost 1.76 release.


olk:
ty - already fixed

PR #169, opened 2021-05-09
Update: minor

snxx-lppxx:


olk:
sorry - I don't like reformatting the code. especially the curly brackets for single line if-statements are set by intention.

PR #174, opened 2021-06-12
Cease dependence on Thread

Kojoley:
On C++11 static local variables are initialized in thread-safe manner, but even on C++03 it should not be a problem because in our case variables are of trivial types, which means double initialization is not an issue, and they are initialized with the same value in every thread.


pdimov:
Looks good. If accepted, be sure to delete the dependency from `CMakeLists.txt` as well: https://github.com/boostorg/context/blob/131ec9db2416a6840b56d62e1e39262274ea31b5/CMakeLists.txt#L177-L178

olk:
But because of the dependency of boost.asio to boost.coroutine boost.context must be compile able with C++03 compilers too!
So  static local variables are not initialized in a thread safe manner.

Kojoley:
> But because of the dependency of boost.asio to boost.coroutine boost.context must be compile able with C++03 compilers too!
> So static local variables are not initialized in a thread safe manner and `call_once()` is still required.

Did you read the PR description? I'll quote:

> but even on C++03 it should not be a problem because in our case variables are of trivial types, which means double initialization is not an issue, and they are initialized with the same value in every thread.

olk:
OK, ty

olk:
ty

Kojoley:
np, I have almost identical PR against Boost.Coroutine library https://github.com/boostorg/coroutine/pull/58, I would appreciate a merge there too.

PR #185, opened 2021-08-11
Fix 64-bit XCOFF assembly

### LINKED ISSUE #180: XCOFF version of context assembly doesn't match ABI?
I'm porting boost::context's XCOFF (you know, AIX and its ABI) assembly for PHP's own fibers (see php/php-src#7338; PHP's own), but I noticed some oddities with the assembly. (This isn't essential because of a `ucontext_t` fallback, but I got curious as to why.)

* The xcoff/ppc64 version of `make_fcontext` truncates values to 32-bit. This seems to be fixable with
* The xcoff/ppc64 versions of make/jump seem to not be using the AIX ABI (!). Notably, they don't seem aware of function descriptors; the ppc64 _CALL_ELF==1 code in the ELF version handles this, because the AIX ABI is mostly identical to ELFv1. Likewise, jump seems to restore r13 instead of r2 (TOC) instead; IBM's own docs say this is reserved, and the ELF code just restores r2 for v1 and nothing for v2.

These seem so fundamental (and attached to the build system), so having them not work at all is a bit confusing. Especially the fact they're for the IBM and not GNU assembler, which has syntax differences.

NattyNarwhal:
It should almost exactly match the ELFv1 ABI code now, except with IBM assembler syntax.

Tested with [PHP fibres](https://github.com/php/php-src/pull/7338). Fixes #180

### DIFF
--- a/src/asm/jump_ppc64_sysv_xcoff_gas.S
+++ b/src/asm/jump_ppc64_sysv_xcoff_gas.S
@@ -12,7 +12,7 @@
     # reserve space on stack
     subi  1, 1, 184
 
-    std  13, 0(1)  # save R13
+    std  2, 0(1)  # save TOC
     std  14, 8(1)  # save R14
     std  15, 16(1)  # save R15
     std  16, 24(1)  # save R16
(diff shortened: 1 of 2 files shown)


olk:
ty

PR #188, opened 2021-09-24
restore work with TLS[__PTK_LIBC_DYLD_Unwind_SjLj_Key] on arm aapcs macho

apolukhin:
I've found this fix in our codebase. Not sure how it works, but my guess is following:

`__PTK_LIBC_DYLD_Unwind_SjLj_Key` is defined in https://opensource.apple.com/source/libpthread/libpthread-137.1.1/private/tsd_private.h as `#define __PTK_LIBC_DYLD_Unwind_SjLj_Key	18`. Multiplying it on a size of a pointer gives `72`.

P.S.: it already was `72` before https://github.com/boostorg/context/commit/763f28542d3c2931ead6eef1409b113155e9a1c9

### DIFF
--- a/src/asm/jump_arm_aapcs_macho_gas.S
+++ b/src/asm/jump_arm_aapcs_macho_gas.S
@@ -52,7 +52,7 @@ _jump_fcontext:
     bic  v2, v2, #3
 
     @ load TLS[__PTK_LIBC_DYLD_Unwind_SjLj_Key]
-    ldr  v1, [v2, #8]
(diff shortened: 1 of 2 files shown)


olk:
So MacOS uses setjump/long jump for exception handling implementation?

apolukhin:
> So MacOS uses setjump/long jump for exception handling implementation?

Not sure. Just showing a patch and my guesses why it does a proper change

olk:
ty - could you check branch develop, please

Issue #130, opened 2019-12-15
Support fcontext_t for Windows on ARM64


yurybura:

olk:
you are welcome to provide the necessary assembler code

janisozaur:
I've provided necessary fix across https://github.com/boostorg/context/pull/122 and https://github.com/boostorg/build/pull/502, there's still a patch needed to coroutine, I think it's the final piece. I had it mostly working in naive version that didn't pass one of the tests (I think it was the fcontext one?). Sadly, I may have lost the patch, but will try restoring it and publish it again. While the test failed, the patch still unblocked other work as I believe it could complete boost installation with vcpkg.

olk:
closed because no progress/no assembler provided

PR #201, opened 2022-06-26
Windows arm64 fcontext support

dixyes:
Tested on m1 mac mini win11 vm and win10 at galaxy book S with `..\..\..\b2 abi=aapcs architecture=arm fc`

CMake can build libraries but no cmakefile for tests...

issue: #130


twose:
Ping @olk 😃
We need this to support Windows arm64 🤩.

olk:
ty

PR #205, opened 2022-09-17
Support for stack protection

apolukhin:


apolukhin:
Hm... Looks like something is wrong with tins patch, however it works fine in our code base. I'll try to investigate further, hints would be appreciated

olk:
ty

PR #206, opened 2022-09-19
CI: Test multiple architectures

Kojoley:
Run tests on most of supported platforms via QEMU. There seems to be no distro that supports powerpc, debian jessie packages for powerpc are no longer available.

Some changes I had to make:
- Run segmented stack tests only when compiler has `-fsplit-stack` and targeted libstdc++ (adds Clang, limits on unsupported arches and stdlibs).
- Switch to lightweight tester from Boost.Test. It was failing mingw for some reason, and this also speed up qemu runners at least x2 (ppc64el 28mins vs 11mins).

### DIFF
--- a/.github/workflows/multi.yml
+++ b/.github/workflows/multi.yml
@@ -1,0 +1,142 @@
+#==============================================================================
+#   Copyright (c) 2022 Nikita Kniazev
+#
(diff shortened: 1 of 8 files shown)


olk:
ty

PR #207, opened 2022-09-30
Add shadow stack support for fcontext

PeterYang12:
Shadow stack is part of Intel's Control-Flow Enforcement Technology.

    Whenever a function is called, the return address is pushed onto both
    the regular stack and the shadow stack. When that function returns, the
    return addresses are popped off both stacks and compared; if they fail
    to match, #CP raised.

    Backport this commit from https://github.com/php/php-src/pull/9283
    With this commit, we create shadow stack with syscall map_shadow_stack
    (no.451) for each fiber context and switch the shadow stack accordingly
    during fcontext switch.

Signed-off-by: PeterYang12 <yuhan.yang@intel.com>
Signed-off-by: chen-hu-97 <hu1.chen@intel.com>

### DIFF
--- a/doc/stack.qbk
+++ b/doc/stack.qbk
@@ -372,5 +372,13 @@ Boost.Context headers if stack protection is enabled.
 
 [endsect]
 
+[section:shadow_stack Support for shadow stack protection]
+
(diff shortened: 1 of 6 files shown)


olk:
ty

Issue #184, opened 2021-08-11
Segfault with ucontext on Mac OS with Clang and GCC


francoisk:
All of the examples I tried (examples/fiber/fibonacci, for one) segfault on Mac OS with Clang and GCC if Context is built using ucontext.

The tested configurations are:
- Mac OS 10.15, Clang 11.0
- Mac OS 11, Clang 11.1 and 12
- Mac OS 11, GCC 11.1

These configurations were tested under a different build system (which I ported to myself) but I did test the Boost build on Mac OS 11 with Clang 12 and had the same results. The Fibonacci example produced the correct output on Linux, FreeBSD, and Windows under the ported build.

The build commands I used are:

```
  b2 toolset=clang cxxflags='-std=c++11' --with-context context-impl=ucontext stage
  cd libs/context/examples/fiber
  b2 toolset=clang cxxflags='-std=c++11' context-impl=ucontext
```

olk:
Don' own a MacOS system. Without a stack trace I can't do anything.

francoisk:
Hope this helps. On Mac OS 11 with Clang 12.0:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x10040458e)
  * frame #0: 0x00007fff2013cc94 libsystem_malloc.dylib`tiny_free_list_add_ptr + 122
    frame #1: 0x00007fff2013a3ff libsystem_malloc.dylib`tiny_malloc_from_free_list + 1632
    frame #2: 0x00007fff2013981d libsystem_malloc.dylib`tiny_malloc_should_clear + 233
    frame #3: 0x00007fff20138727 libsystem_malloc.dylib`szone_malloc_should_clear + 66
    frame #4: 0x00007fff20151fe5 libsystem_malloc.dylib`_malloc_zone_malloc + 118
    frame #5: 0x00007fff2032940d libdyld.dylib`_tlv_atexit + 172
    frame #6: 0x0000000100006a18 fibonacci`boost::context::detail::fiber_activation_record::current() at fiber.cpp:48:5
    frame #7: 0x00000001000049d9 fibonacci`boost::context::detail::fiber_activation_record::resume(this=0x0000000100170f00) at fiber_ucontext.hpp:131:10
    frame #8: 0x0000000100004755 fibonacci`boost::context::fiber::resume(this=0x00007ffeefbff900) && at fiber_ucontext.hpp:471:83
[...]

Issue #198, opened 2022-05-05
heap-buffer-overflow switching contexts with -DBOOST_USE_UCONTEXT and -DBOOST_USE_ASAN


elfprince13:
On boost 1.78, built with context-impl=ucontext.

Reproducer with log:

```console
ubuntu@host:~$ git clone --recursive --branch crash-reproducer-2022.05.05 https://github.com/Geopipe/Cxx-Bidirectional-Coroutines.git
Cloning into 'Cxx-Bidirectional-Coroutines'...
remote: Enumerating objects: 149, done.
remote: Counting objects: 100% (82/82), done.
remote: Compressing objects: 100% (62/62), done.
remote: Total 149 (delta 32), reused 62 (delta 16), pack-reused 67
Receiving objects: 100% (149/149), 48.66 KiB | 3.74 MiB/s, done.
Resolving deltas: 100% (55/55), done.
Submodule 'deps/Functional-Cxx' (ssh://git@github.com/Geopipe/Functional-Cxx.git) registered for path 'deps/Functional-Cxx'
Cloning into '/home/ubuntu/Cxx-Bidirectional-Coroutines/deps/Functional-Cxx'...
remote: Enumerating objects: 68, done.        
remote: Counting objects: 100% (68/68), done.        
remote: Compressing objects: 100% (44/44), done.        
remote: Total 68 (delta 24), reused 56 (delta 16), pack-reused 0        
Receiving objects: 100% (68/68), 33.01 KiB | 11.00 MiB/s, done.
Resolving deltas: 100% (24/24), done.
Submodule path 'deps/Functional-Cxx': checked out 'c79c1fa55376275349afa618db77db88e4088b53'
ubuntu@host:~$ cd Cxx-Bidirectional-Coroutines/
ubuntu@host:~/Cxx-Bidirectional-Coroutines$ mkdir build
ubuntu@host:~/Cxx-Bidirectional-Coroutines$ cd build/
ubuntu@host:~/Cxx-Bidirectional-Coroutines/build$ cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="-fsanitize=address -fsanitize=undefined -fsanitize-recover=all  -DBOOST_USE_ASAN -DBOOST_USE_UCONTEXT" ..
-- The C compiler identification is Clang 12.0.0
-- The CXX compiler identification is Clang 12.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
[...]

PR #213, opened 2022-12-16
fix ucontext for MacOS

eduardvoronkin:
1. MacOS version of ::makecontext (tested on m1 Mac Ventura 13.0.1) is able to pass only 4-byte parameters to context function
2. define _XOPEN_SOURCE before inclusion of predef.h

https://github.com/boostorg/context/issues/184

### DIFF
--- a/include/boost/context/continuation_ucontext.hpp
+++ b/include/boost/context/continuation_ucontext.hpp
@@ -7,7 +7,7 @@
 #ifndef BOOST_CONTEXT_CONTINUATION_H
 #define BOOST_CONTEXT_CONTINUATION_H
 
-#include <boost/predef.h>
+#include <boost/predef/os.h>
 #if BOOST_OS_MACOS
 #define _XOPEN_SOURCE 600
 #endif
@@ -16,6 +16,7 @@ extern "C" {
 #include <ucontext.h>
 }
 
+#include <boost/predef.h>
 #include <boost/context/detail/config.hpp>
 
 #include <algorithm>
@@ -60,12 +61,19 @@ namespace detail {
 // tampoline function
 // entered if the execution context
 // is resumed for the first time
-template< typename Record >
-static void entry_func( void * data) noexcept {
-    Record * record = static_cast< Record * >( data);
-    BOOST_ASSERT( nullptr != record);
-    // start execution of toplevel context-function
-    record->run();
+template <typename Record>
+#ifdef BOOST_OS_MACOS
+static void entry_func(std::uint32_t data_high,
+                       std::uint32_t data_low) noexcept {
+  auto data =
+      reinterpret_cast<void *>(std::uint64_t(data_high) << 32 | data_low);
+#else
+static void entry_func(void *data) noexcept {
+#endif
+  Record *record = static_cast<Record *>(data);
+  BOOST_ASSERT(nullptr != record);
+  // start execution of toplevel context-function
+  record->run();
 }
 
 struct BOOST_CONTEXT_DECL activation_record {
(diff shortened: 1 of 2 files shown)


elfprince13:
So sounds like this likely _doesn't_ solve #198?

eduardvoronkin:
> So sounds like this likely _doesn't_ solve #198?

This should solve this. At least, we got boost asio/context working with address sanitizer. Another tricky thing is that you need to keep track of is the _XOPEN_SOURCE macro. I fixed this issue in context, but the same issue remains in boost.asio and actually this is MacOS issue. There are some "bad" headers that lead to inclusion of /sys/_types/_ucontext.h (the file with definition of ucontext struct) WITHOUT this macro. So, as you see from definition 
```cpp
_STRUCT_UCONTEXT
{
	int                     uc_onstack;
	__darwin_sigset_t       uc_sigmask;     /* signal mask used by this context */
	_STRUCT_SIGALTSTACK     uc_stack;       /* stack used by this context */
	_STRUCT_UCONTEXT        *uc_link;       /* pointer to resuming context */
	__darwin_size_t         uc_mcsize;      /* size of the machine context passed in */
	_STRUCT_MCONTEXT        *uc_mcontext;   /* pointer to machine specific context */
#ifdef _XOPEN_SOURCE
	_STRUCT_MCONTEXT        __mcontext_data;
#endif /* _XOPEN_SOURCE */
};
```
if we include this WITHOUT _XOPEN_SOURCE, the struct will lack one of its field. But then, lately, we include sys/ucontext.h, which expects this macro to be defined, and it's actually defined. And then some weird issues happening (getcontext does memory overrun).

One of examples of such bad include sequence:
```
In file included from /myproj/boost-src/boost/asio/append.hpp:24:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/c++/v1/tuple:154:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/c++/v1/__functional_base:22:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/c++/v1/exception:83:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/c++/v1/cstdlib:85:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/c++/v1/stdlib.h:93:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/stdlib.h:66:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/sys/wait.h:109:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/sys/signal.h:151:
```
and then signal.h on line 151 includes `#include <sys/_types/_ucontext.h>`.

To summarize: 
1. Need fix for entry_func, bc makecontext does not work properly (at lease, on my MacOS EVEN with _XOPEN_SOURCE).
2. Need to make sure that _XOPEN_SOURCE is defined for sources that are using boost.context directly or indirectly. If you building with CMake, just add ```target_compile_definitions(boost_context PUBLIC _XOPEN_SOURCE)```

rwsfisker:
> So sounds like this likely _doesn't_ solve #198?

It does:
```

# clone boost, build boost_context
git clone --recurse-submodules -j8 https://github.com/boostorg/boost.git
cd boost
boost_root=`pwd`
./bootstrap.sh --with-libraries=context
./b2 -a cxxflags='-std=c++17' context-impl=ucontext stage

# clone / build crash reproducer
git clone --recursive --branch crash-reproducer-2022.05.05 https://github.com/Geopipe/Cxx-Bidirectional-Coroutines.git
mkdir Cxx-Bidirectional-Coroutines/build
cd Cxx-Bidirectional-Coroutines/build
cmake \
 -DCMAKE_BUILD_TYPE=Debug \
 -DCMAKE_CXX_FLAGS="-fsanitize=address -fsanitize=undefined -fsanitize-recover=all -DBOOST_USE_ASAN -DBOOST_USE_UCONTEXT" \
 -DCMAKE_PREFIX_PATH=${boost_root}/stage/lib/cmake/Boost-1.82.0/ \
 -S.. -B.
cmake --build .

# it crashes
./example/testcoro # crashes

# apply the patch / rebuild
test_root=`pwd`
cd ${boost_root}/libs/context
curl https://github.com/boostorg/context/commit/cc106be672f54b71842aa6322fe45bf60ed1d3cc.diff | git apply -
cd ${boost_root}
./b2 -a cxxflags='-std=c++17' context-impl=ucontext stage

# build it again, it works!
cd ${test_root}
cmake --build .
% ./example/testcoro
Fibs
0
1
1
2
3
5
8
13
21
34
RBC
1
1
1
2
3
3
4
Noise
Moo/0
Moo/1
Moo/2
Moo/3
Moo/4
Moo/5
Moo/6

```

rwsfisker:
Verified Mac OSX 13, x86_64

olk:
ty

Issue #211, opened 2022-11-24
We need some help to finally fix Darwin PowerPC assembler so that not just build succeeds but also tests pass


barracuda156:
In a couple of recent PRs I have made some fixes to Darwin PPC assembler here (nothing big, just fixing which was actually breaking the build), which fixed _building_ of `context` and made it _somewhat_ functional (in a sense that certain dependencies that require `context` now build successfully, `folly` and friends being the case).
However, tests do not pass: https://github.com/macports/macports-ports/pull/16407#issuecomment-1304974991

From earlier PRs by @kernigh and @DaoWen for PPC ELF it looks like the issue is quite non-trivial and requires a thorough understanding of both what Boost is doing and the _magic code_ :)
Unfortunately, my knowledge of PPC assembler is rather limited and does not suffice here. I am not going to drop the case, since we really want this fixed properly, but any advice or help will be greatly appreciated.

_This has been broken apparently forever – can we fix it now?_ :)

(Considering a recent case with Boost shared pointer (https://github.com/boostorg/smart_ptr/issues/105), I cannot be entirely sure that it is assembler which is to blame and not again some wrong alignment, but probably it is assembler, given that it was also broken for *BSD.)

P. S. For the record, alternative version of PPC code for `context` from the following repo also fails, at least with Fibonacci test: https://github.com/twlostow/libcontext

P. P. S. @iains I am aware you are very busy atm, so please feel free to completely ignore the issue. But still tagging you as the person who understands Darwin magic code best.

olk:
Seams to be a NULL pointer is dereferenced - this might happen because of the calling convention is violated.
I don't own a PPC64(?) system and I'm too busy atm.
You could debug a simple test app (simple jump) by stepping through the assembler instructions and compare the content in teh registers with what the calling convention requires.

kernigh:
My PPC32 hardware is running OpenBSD, but I have a QEMU running Darwin. I can't compile Boost for Darwin, because my gcc 3.3 can't do C++11, but I can compile *ppc32_sysv_macho*.S and call them from C.

I found a problem: `make_fcontext(_, _, there)` fails to pass the transfer_t there. It is passing r3 = address of transfer_t, but there wants r3:r4 = transfer_t. You can jump there, but there can't jump to its transfer_t's fctx; jump_fcontext will crash.

This diff seems to fix make_context, but ontop_context might still be broken.

```
[...]

PR #215, opened 2022-12-27
Fixes for Darwin PPC32

evanmiller:
This PR consolidates some of @kernigh's work over in #211. It successfully compiles and runs the Fibonacci test, without a crash, on PPC32 hardware running Darwin 8:

```
$ ./a.out
0 1 1 2 3 5 8 13 21 34
main: done
$ uname -r
8.11.0
```

For `make_fcontext`, use the diff provided here:

https://github.com/boostorg/context/issues/211#issuecomment-1328154669

For `ontop_context`, adapt the Linux PPC32 fixes from here:

https://github.com/boostorg/context/commit/df8fb6b5289492914dd2af479e40e9737399fd21


barracuda156:
@evanmiller Do you think we should first ensure that `fiber` does not fail with these changes? Given that is was building before with not fully fixed `context`, it presumably should build with the properly fixed one.

evanmiller:
> @evanmiller Do you think we should first ensure that `fiber` does not fail with these changes? Given that is was building before with not fully fixed `context`, it presumably should build with the properly fixed one.

@barracuda156 Which compilers / Boost versions were successfully building Fiber on PPC? I encountered an issue with GCC7 / Boost 1.81 but I can't comment on any other combinations.

barracuda156:
> > @evanmiller Do you think we should first ensure that `fiber` does not fail with these changes? Given that is was building before with not fully fixed `context`, it presumably should build with the properly fixed one.
> 
> @barracuda156 Which compilers / Boost versions were successfully building Fiber on PPC? I encountered an issue with GCC7 / Boost 1.81 but I can't comment on any other combinations.

Well, everything was building with my earlier minimal fixes to `context` (which are not sufficient, as we know now).

evanmiller:
@barracuda156 If you were able to build everything before, why don't you try building Fiber with these proposed changes and report back?

barracuda156:
> @barracuda156 If you were able to build everything before, why don't you try building Fiber with these proposed changes and report back?

@evanmiller I will do it soonish.

olk:
ty

barracuda156:
Now we have `ppc64` left to fix. And possibly `fiber` for both.

barracuda156:
@evanmiller Could you remind, what prevented us from fixing `ppc64` along?

Issue #210, opened 2022-11-11
i386 sysv elf assembly for make_fcontext seems to leave stack misaligned


alexfusco5:
In [this file](https://github.com/boostorg/context/blob/develop/src/asm/make_i386_sysv_elf_gas.S#L44), it seems to be leaving the resulting context stack misaligned.  I'm fairly rusty with x86 assembly so I don't have a great handle on why this is the case, but it seems that the offset on line 44 should be 0x34 rather than 0x2c.  The stack otherwise appears to be aligned improperly

olk:
ty

PR #218, opened 2023-01-25
Make sure the i386 sysv stack is aligned

nielsdos:
SysV ABI requires a stack alignment of 16 bytes. Currently, for i386 with SysV ABI, the trampoline function is entered with an unaligned stack. This causes problems for the context-function that is jumped to as its stack is also unaligned. This causes a crash for our use-case because the context function contains an SSE instruction which reads from the stack. The SSE instruction requires the correct alignment. Fix it by changing the 0x2c offset to 0x30, such that the stack remains aligned.

Related issue in this repo: #210 

For reference:
https://github.com/php/php-src/pull/10407#issuecomment-1404180877 and https://github.com/php/php-src/issues/10398


olk:
ty

PR #224, opened 2023-05-11
Store entry point arguments in fibre's own stack (s390x)

ltrk2:
In the current implementation fibres spawned by other fibres or threads would have the argument of their entry point (`void fiber_entry( transfer_t t)`) be allocated on the stack of their parent, which would then cause segmentation fault if the former outlived the latter.

The x86-64 implementation (`x86_64_sysv_elf` is the combination I tested) supports this use-case, so this change means to achieve feature parity for the s390x implementation in this regard by storing the aforementioned argument in the fibre's own stack.


olk:
ty

PR #225, opened 2023-05-15
Jamfile: sort out macOS ppc asm_sources

barracuda156:
@olk Bring consistency there, add `gcc`. (In fact, presently `clang` is broken on macOS ppc, but let it be there, maybe it gets fixed some day.)


olk:
ty

PR #226, opened 2023-05-15
Improve ABI conformance of s390x fcontext

ltrk2:
I implemented review comments by @Andreas-Krebbel from #224. I ran the tests using `b2` (`../../../b2 -a`) and everything seems to be passing.


olk:
ty

PR #236, opened 2023-08-18
Fix check for BOOST_OS_MACOS

iv-m:
BOOST_OS_MACOS is always defined; when we're not compiling for Mac OS, it's just defined as zero. So, the correct check is `#if BOOST_OS_MACOS`, because `#ifdef BOOST_OS_MACOS` is true on any operating system.

Fixes: cc106be672f54b71842aa6322fe45bf60ed1d3cc
Reported-by: Ilya Kurdyukov <ilyakurdyukov@altlinux.org>


olk:
ty

PR #240, opened 2023-09-06
CMake: Fix assembler `--defsym` format issue (qcc/qnx)

### LINKED ISSUE #239: CMake builds fails on QNX (qcc): Incorrect assembler invocation
While trying to build Boost for QNX with CMake, I faced the issue that `-A,--defsym` assmbler flags were not accepted by qcc.
Only when I replaced all defines without assignment.

```cmake
target_compile_definitions(boost_context VAR)
```
into

```cmake
target_compile_definitions(boost_context VAR=)
```

would the build succeed. I did not face the issue while building with `b2`.
Would it be possible to either really apply this approach and/or watch more closely which definitions are passed to the assembler?

Felix-El:
Adjust compile definitions to adhere to KEY=VALUE (where VALUE is empty string) so it builds with qcc. Closes #239

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,23 +214,26 @@ target_link_libraries(boost_context
     Boost::smart_ptr
 )
 
+# ATTENTION. Some assemblers are picky to get --defsym,KEY=VALUE pairs where '=' is non-optional!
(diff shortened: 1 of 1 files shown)


olk:
ty

PR #247, opened 2024-01-23
Fix x32 build on x86-64

SoapGentoo:
@olk we have been carrying this patch for 10 years, I'd like to get it upstreamed so we can drop it from Gentoo


olk:
ty

PR #248, opened 2024-02-04
fix solaris/illumos build on make/jump assembly directives.

devnexen:
on those platforms, albeit being 64 bits, they produce 32 bits binaries by default.


olk:
ty

PR #250, opened 2024-02-13
Add the BOOST_CONTEXT_USE_MAP_STACK logic to test_fcontext.cpp

cjeker:
Without this test_fcontext.cpp fails on OpenBSD with a not-on-stack OS error.

This is just bringing the same logic from the lib code to this test code.


olk:
ty

PR #262, opened 2024-06-18
Convert ARM64 armasm to armclang

Liryna:
The armasm legacy assembler is deprecated. The code was migrated to the GNU syntax (used by armclang).
https://developer.arm.com/documentation/100068/0621/Migrating-from-armasm-to-the-armclang-Integrated-Assembler

Split this into two commit to keep the history and diff clean

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -253,6 +253,18 @@ alias asm_sources
      <toolset>msvc
    ;
 
+# ARM64/AAPCS/PE
+alias asm_sources
+   : asm/make_arm64_aapcs_pe_armclang.asm
+     asm/jump_arm64_aapcs_pe_armclang.asm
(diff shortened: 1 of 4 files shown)


dixyes:
Does Microsoft provide armclang in the default installation of Visual Studio Code?

These codes are used in many projects. If armasm support is removed, they may require users to install additional dependencies (and a build system modification).

dixyes:
BTW the armasm in file name means [ms armasm64](https://learn.microsoft.com/en-us/cpp/assembler/arm/arm-assembler-command-line-reference?view=msvc-170), not ARM armasm, they are not deprecated yet.

So I think keeping these armasm64 sources, adding new armclang sources/b2 configurations then wait for ms providing armclang is better

Liryna:
Correct, I couldn't find msvc support armclang for now 😢 
Therefore I kept the existing one and added the converted armclang only

PR #266, opened 2024-09-06
Trim redundant trailing whitespace

petk:
Hello,

this trims redundant trailing whitespace characters in all files in this repository. I was thinking to only change the src/asm files but since the diff in other files seem pretty simple this changes all files in the repo. Reason I'm doing this is that it is a bit easier to integrate in other repositories that use various code-style tools. 

Thank you.


olk:
ty

PR #268, opened 2024-09-12
fix for for: visual studio missing buildtool peroperty for asm files

Dargun:
There is a problem on visual studio to compile Boost::context with BOOST_CONTEXT_IMPLEMENTATION = "fcontext".
For all asm files are no build tool defined so they will not be compiled during build of context lib and functions "jump_fcontext, make_fcontext and ontop_fcontext" are no compiled.
The fix is to pass the defined ASM_LANGUAGE on each asm file as PROPERTY.


olk:
ty

Issue #269, opened 2024-09-12
`__NR_map_shadow_stack` redefinition


Lastique:
When compiling Boost.Fiber on Kubuntu 24.04 (kernel version 6.8.0-44, glibc 2.39-0ubuntu8.3) I get these warnings:

```
In file included from ./boost/context/fiber.hpp:12,
                 from ./boost/fiber/context.hpp:28,
                 from libs/fiber/src/algo/algorithm.cpp:9:
./boost/context/fiber_fcontext.hpp:52: warning: "__NR_map_shadow_stack" redefined
   52 | #  define __NR_map_shadow_stack 451
      | 
In file included from /usr/include/x86_64-linux-gnu/asm/unistd.h:20,
                 from /usr/include/x86_64-linux-gnu/sys/syscall.h:24,
                 from /usr/include/syscall.h:1,
                 from /usr/include/c++/13/bits/atomic_wait.h:46,
                 from /usr/include/c++/13/bits/atomic_base.h:42,
                 from /usr/include/c++/13/atomic:41,
                 from ./boost/fiber/algo/algorithm.hpp:9,
                 from libs/fiber/src/algo/algorithm.cpp:7:
/usr/include/x86_64-linux-gnu/asm/unistd_64.h:368: note: this is the location of the previous definition
  368 | #define __NR_map_shadow_stack 453
      | 
```

I think, the code should check if `__NR_map_shadow_stack` is defined and use the existing define, if possible.

I should say that `__NR_map_shadow_stack` is defined to 453 in both `unistd_64.h` and `unistd_32.h` on my machine. I'm not sure if the value 451 is correct to begin with.

anarthal:
Could I request this to get merged to master, please?

Issue #263, opened 2024-07-08
gcc-14: map_shadow_stack incorrect syscall number


anarthal:
System: Ubuntu 24.04 (as per the `ubuntu:24.04` Docker image)
Compiler: g++-14 (as per `apt install g++-14` in the aforementioned image)
uname -a: `Linux 2b064e2a8785 6.5.0-41-generic #41~22.04.2-Ubuntu SMP PREEMPT_DYNAMIC Mon Jun  3 11:32:55 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`
glibc version: 2.39
Variant: release
Library version: Boost develop branch

I'm getting a warning about a redefinition of `__NR_map_shadow_stack`. It's not causing any issues, in part because I'm only indirectly including the file (via Asio's `spawn.hpp`). It looks like `<boost/context/fiber.hpp>`  is defining `__NR_map_shadow_stack` to 451, but the actual syscall number seems to be 453. This only happens in C++23, where `<memory>` includes all the syscall numbers.

For me it's just an annoying warning, but it seems it may point to a real bug. I don't know enough to provide further diagnostics.

Please note that this doesn't happen in the Godbolt environment, which uses an older glibc version. The ubuntu 24.04 gcc-14 seems to have `__CET__` defined by default.

Repro instructions:

```
$> cd $BOOST_ROOT
$> cat libs/mysql/repro/Jamfile
exe repro :
    repro.cpp
    /boost/context//boost_context ;

$> cat libs/mysql/repro/repro.cpp
#include <boost/context/fiber.hpp>

int main() {}

$> ./b2 toolset=gcc-14 cxxstd=23 libs/mysql/repro warnings-as-errors=on
Performing configuration checks

    - default address-model    : 64-bit (cached) [1]
    - default architecture     : x86 (cached) [1]
    - symlinks supported       : yes (cached)

[1] gcc-14
...patience...
...found 639 targets...
...updating 9 targets...
gcc.compile.c++ bin.v2/libs/mysql/repro/gcc-14/debug/x86_64/cxxstd-23-iso/threading-multi/visibility-hidden/repro.o
In file included from ./boost/context/fiber.hpp:12,
                 from libs/mysql/repro/repro.cpp:1:
./boost/context/fiber_fcontext.hpp:52:11: error: "__NR_map_shadow_stack" redefined [-Werror]
   52 | #  define __NR_map_shadow_stack 451
      |           ^~~~~~~~~~~~~~~~~~~~~
[...]

PR #270, opened 2024-09-12
Fix shadowstack syscall number redefinition

Lastique:
`__NR_map_shadow_stack` is defined to 453 (not 451) on Ubuntu 24.04 with Linux kernel 6.8.0-44, gcc 13.2.0 and glibc 2.39. This causes warnings about macro redefinition when Boost.Fiber is compiled.
    
Change the syscall number to 453 and only define the macro if it has not been defined already. Also include `unistd.h` for `syscall()` and `__NR_*` constants.

Also convert tabs to spaces and trim trailing spaces.

Fixes https://github.com/boostorg/context/issues/263.
Fixes https://github.com/boostorg/context/issues/269.


olk:
ty

PR #273, opened 2024-10-14
Convert make_fcontext etc into true C++ symbols

rdoeffinger:
Also mark the non-mangled fcontext asm function symbols as hidden visibility
These functions should not be exported as dynamic
symbols by boost, only the namespaced C++ symbols
introduced by this patch should.

Note: I did not test the cmake change - I hope someone can do that more easily than me figuring it out.
Apologies if the copyright on the new file seems a bit overkill given there is hardly any new code, but it seemed the most proper way of doing it.


olk:
ty

PR #275, opened 2024-11-11
Define assembler files for ARM64 + Windows + GCC combination

vejbomar:
Adds a definition which assembler files should be used for ARM64 + Windows + GCC combination. (This is needed for a new target `aarch64-w64-mingw32` for MinGW which is being developed here: https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build)

It is enough to use the existing assembler for Clang on Arm64 (introduced here https://github.com/boostorg/context/pull/262) as they both use the GNU syntax.

Also the first commit fixes the assembler extensions for Clang.

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -204,9 +204,9 @@ alias asm_sources
 
 # ARM64/AAPCS/PE
 alias asm_sources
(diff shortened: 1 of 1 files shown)


eukarpov:
@olk Could you please review this change?

olk:
ty

PR #276, opened 2024-12-15
Support building assembly files for mingw-w64 on arm64 with CMake.

MehdiChinoune:
We used to build boost-context on MSYS2/clangarm64 (https://github.com/msys2/MINGW-packages/commit/da4f4c37).


olk:
ty

PR #281, opened 2025-01-13
On libstdc++, fiber_fcontext.hpp sets fiber-specific exception state.

nat-goodspeed:
This makes use of the libstdc++ published entry point `__cxa_get_globals()`, which retrieves a pointer to the `__cxa_eh_globals` struct. Saving and restoring that struct on every context switch ensures fiber-specific exception behavior.


olk:
ty

PR #234, opened 2023-08-17
Fix ABI detection for empty 'os.platform'

iv-m:
`in` operator in bjam always returns true if its first argument has no elements[1]. This means that if `os.platform` is empty (not detected), the construction introduced in commit d039c8e4da79dbc76481236249e22892bfdde047 sets ABI to `aapcs` on all platforms where `os.platform` is empty, including, e.g. riscv64, and breaks build there.

This commit refactors the condition to use '=' operator, to make sure that when `os.platform` is empty we get the default ABI value, and thus fixes build on riscv64.

[1] https://www.boost.org/doc/libs/1_83_0/tools/build/doc/html/index.html#jam.language.flow_of_control

Fixes: d039c8e4da79dbc76481236249e22892bfdde047

PR #283, opened 2025-01-23
Fix ABI detection for empty 'os.platform'

iv-m:
`in` operator in bjam always returns true if its first argument has no elements[1]. This means that if `os.platform` is empty (not detected), the construction introduced in commit 9084161b8778c18ebce7514f886f0f22f9bff490 sets ABI to `aapcs` on all platforms where `os.platform` is empty, including, e.g. loongarch64, and breaks build there.

This commit refactors the condition to use '=' operator, to make sure that when `os.platform` is empty we get the default ABI value, and thus fixes build on loongarch64.

See also 819c2d6423b4e0f55c5f69e334bf81570e82f68f for similar fix that was somehow lost in refactorings. This time a comment is added near the condition in hope that future edits will not reintroduce the issue.

[1] https://www.boost.org/doc/libs/1_83_0/tools/build/doc/html/index.html#jam.language.flow_of_control

Fixes: 9084161b8778c18ebce7514f886f0f22f9bff490

### DIFF
--- a/boost-context-features.jam
+++ b/boost-context-features.jam
@@ -37,9 +37,14 @@ feature.set-default binary-format : [ default_binary_format ] ;
 local rule default_abi ( )
 {
     local tmp = sysv ;
+
+    # Avoid using "in" operator here: it returns true if its left
+    # part is empty, which happens e.g. with os.platform on
+    # some uncommon architectures.
     if [ os.name ] = "NT" { tmp = ms ; }
     else if [ os.name ] = "CYGWIN" { tmp = ms ; }
-    else if [ os.platform ] in "ARM" "ARM64" { tmp = aapcs ; }
+    else if [ os.platform ] = "ARM" { tmp = aapcs ; }
+    else if [ os.platform ] = "ARM64" { tmp = aapcs ; }
     else if [ os.platform ] = "MIPS32" { tmp = o32 ; }
     else if [ os.platform ] = "MIPS64" { tmp = n64 ; }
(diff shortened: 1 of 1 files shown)


iv-m:
I faced this issue when building boost 1.87.0 on loongarch64. The b2 output and fix are similar to #234:

`error: No best alternative for /usr/src/RPM/BUILD/boost-1.87.0/libs/context/build/asm_sources with <abi>aapcs <address-model>64 <architecture>loongarch <asynch-exceptions>off <binary-format>elf <boost.beast.allow-deprecated>on <boost.beast.separate-compilation>on <boost.cobalt.executor>any_io_executor <boost.cobalt.pmr>std <boost.process.fs>boost <context-impl>fcontext <coverage>off <debug-symbols>off <exception-handling>on <extern-c-nothrow>off <inlining>full <known-warnings>hide <link>shared <optimization>off <os>LINUX <pch>off <preserve-test-targets>on <profiling>off <python-debugging>off <python>3.12 <relevant>abi <relevant>address-model <relevant>architecture <relevant>binary-format <relevant>toolset <rtti>on <runtime-debugging>off <runtime-link>shared <stdlib>native <strip>off <target-os>linux <testing.execute>on <threadapi>pthread <threading>multi <toolset-gcc:version>14 <toolset>gcc <variant>release <vectorize>off <visibility>hidden <warnings-as-errors>off <warnings>on <x-deduced-platform>loongarch_64`

olk:
ty

PR #291, opened 2025-02-17
fix tls stack protector on x86_64

CBBBrother:
Hello! I find mistake at code under BOOST_CONTEXT_TLS_STACK_PROTECTOR for x86_64.


olk:
ty

PR #305, opened 2025-05-16
fix build of ontop_sparc64_sysv_elf_gas.S

tkoecker:
Fix the symbol names used for .size.

It seems that this line was copy pasted from the "jump" implementation without adjusting the symbol name.


olk:
ty

Issue #308, opened 2025-07-23
AArch64: add BTI hint to *fcontext trampolines - fixes BTI crash on Graviton4


zaydr-al:
### Summary
When Boost.Context is built with **PAC + BTI hardening** (``-mbranch-protection=standard``), the first indirect jump into any trampoline raises **SIGILL** on BTI-enforcing CPUs (Graviton 4).
Cause: the trampolines are missing the mandatory **btic** hint.

LLVM is currently fixing the analogous omission for compiler-generated landing pads in [llvm.llvm-project #149267](https://github.com/llvm/llvm-project/issues/149267) ([PR llvm/llvm-project#149680](https://github.com/llvm/llvm-project/pull/149680)).
Hand-written assembly still needs the hint.

### Minimal repro (Amazon Linux 2023, Boost develop, Clang 15):
```
#1. clone Boost super-repo
git clone --branch develop --recursive https://github.com/boostorg/boost.git boost
cd boost

#2. build Boost.Context only, with PAC+BTI
./bootstrap.sh >/dev/null
./b2 -j$"(nproc)" -q --with-context toolset=clang \
        cxxflags="-O2 -fPIC -mbranch-protection=standard" \
        asmflags="-O2 -fPIC -mbranch-protection=standard" \
        linkflags="-Wl,-z,force-bti" \
        install --prefix=stage

#3. small test that throws inside a fiber
cat > test.cpp <<'CPP'
#include <boost/context/fiber.hpp>
#include <iostream>
int main(){
    boost::context::fiber f([](boost::context::fiber&& s){
        try { throw 1; } catch(int){ std::cout<<"caught\n"; }
        return std::move(s);
    }); 
    f = std::move(f).resume();
}
CPP

clang++ -target aarch64-linux-gnu -mbranch-protection=standard \
        -I$(pwd)/stage/include \
        test.cpp -Lstage/lib -lboost_context \
        -Wl,-z,force-bti \
        -Wl,-rpath,$ORIGIN/../lib -Wl,-rpath,$(pwd)/stage/lib \
        -o test_fiber

# 4. run on Arm host that enforces BTI (Graviton 4)
./test_fiber

```

### Proposed Fix
Insert at the top of each AArch64 trampoline:
```
#if defined(__ARM_FEATURE_BTI_DEFAULT) && (__ARM_FEATURE_BTI_DEFAULT == 1)
/* bti c */
      hint        #34 
#endif
```
### Files
```
[...]

PR #312, opened 2025-08-15
[AArch64][BTI] Add BTI hint + GNU property to fcontext trampolines 

zaydr-al:
**Add BTI hint + GNU property to fcontext trampolines** 

The AArch64 fcontext trampolines (jump_fcontext, make_fcontext, ontop_fcontext) are indirect-entry-points. On BTI-enforcing systems they must begin with a BTI or the first resume can trap with SIGILL.

Insert `bti c` (hint #34) at each entry under `__ARM_FEATURE_BTI_DEFAULT`, and emit `GNU_PROPERTY_AARCH64_FEATURE_1_BTI` from each AArch64 assembly file so linkers map the DSO with PROT_BTI.

Scope: ELF/GAS AArch64 trampolines only; +4 bytes per entry

Fixes [#308](https://github.com/boostorg/context/issues/308)

Notes
- This change is orthogonal to the separate LLVM landing-pad ``bti j`` [issue #149267](https://github.com/llvm/llvm-project/issues/149267); exception paths built with affected Clang toolchains will still require the upstream fix.

### DIFF
--- a/src/asm/jump_arm64_aapcs_elf_gas.S
+++ b/src/asm/jump_arm64_aapcs_elf_gas.S
@@ -53,11 +53,29 @@
 
 .file "jump_arm64_aapcs_elf_gas.S"
 .text
+#if defined(__ARM_FEATURE_BTI_DEFAULT) && (__ARM_FEATURE_BTI_DEFAULT == 1)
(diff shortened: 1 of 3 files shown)


olk:
Did you successfully run the unit tests?

PR #324, opened 2026-01-04
Ensure manage_exception_state uses correct TLS

zhykzhykzhyk:
Solves #323

### DIFF
--- a/include/boost/context/fiber_fcontext.hpp
+++ b/include/boost/context/fiber_fcontext.hpp
@@ -87,10 +87,10 @@ struct __cxa_eh_globals {
 
 class manage_exception_state {
 public:
-    manage_exception_state() {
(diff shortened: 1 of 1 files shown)


CreoValis:
Wouldn't access through a `volatile *` be a better solution here? The "no inline" spec probably works around the issue, but that just seems to push the dtor body to the point where gcc couldn't "see into it" to optimize it away.
I think using a volatile pointer cast would forbid it from even optimizing it away.

Like so:
````cpp
    manage_exception_state() {
        exception_state_ = *(volatile __cxa_eh_globals *)__cxa_get_globals();
    }
    ~manage_exception_state() {
        *(volatile __cxa_eh_globals *)__cxa_get_globals() = exception_state_;
    }
`

PR #333, opened 2026-05-26
Add support for the e2k architecture in the ucontext implementation

arrowd:
Tested by running tests from the `test` subdirectory (except `test_fcontext.cpp`)

CC original author @ilyakurdyukov


olk:
ty

PR #337, opened 2026-07-21
Fix uncaught_exceptions() not accounting for forced_unwind

xokdvium:
Unwound fibers would see std::uncaught_exceptions() == 0, while a forced_unwind exception is in "flight". This goes against the contract of std::uncaught_exceptions() that scope guards rely upon. Failing to report the correct number of uncaught exceptions (especially misreporting zero) will lead to scope guards to misbehave badly and skip running cleanup code which branches on whether the destructor is called during stack unwinding or not.

This is because the "throw" would happen before the destructor is run on the fiber stack being switched to, but the increment would be clobbered by the destructor of manage_exception_state.

I'm not sure what the contract of run ontop_fcontext is wrt to whether the the caller provided function can throw or not, but in my best understanding the forced_unwind mechanism is mostly internal and so is throwing from ontop_fcontext in the switched-to fiber. Thus, I've kept the catch block scoped to detail::forced_unwind.

TODO:

- [x] Add unit tests for this.
- [x] Think about performance implications of the catch block.

### DIFF
--- a/include/boost/context/fiber_fcontext.hpp
+++ b/include/boost/context/fiber_fcontext.hpp
@@ -70,7 +70,9 @@ namespace context {
 namespace detail {
 
 // manage_exception_state is a dummy struct unless we have specific support
-struct manage_exception_state {};
+struct manage_exception_state {
+    void from_forced_unwind() noexcept {}
+};
 
 } // namespace detail
 } // namespace context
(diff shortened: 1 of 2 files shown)


xokdvium:
Ok the new test surfaces the bug under the fcontext backend.

include/boost/context/fiber_fcontext.hpp
```diff
@@ -391,14 +403,19 @@ class fiber {
...
 #else
-                    std::exchange( fctx_, nullptr),
+                        std::exchange( fctx_, nullptr),
 #endif
-                    & p,
-                    detail::fiber_ontop< fiber, decltype(p) >).fctx };
+                        & p,
+                        detail::fiber_ontop< fiber, decltype(p) >).fctx };
+        } catch ( detail::forced_unwind const& ) {
+            exstate.from_forced_unwind();
```
xokdvium, line 416: I guess if an exception is thrown on the switched-to context it would also need to be handled here - and that's probably in-scope. Happy to tackle this (maybe in a follow-up), but I've started by putting out the immediate fire under our feet in nix.

olk:
ty

PR #341, opened 2026-08-19
remove doc artifacts, fix grammar

m-fila:
Remove the html documentation artifacts that can be generated from qbk sources.
Also fix a few wording issues in the documentation.


olk:
ty

PR #342, opened 2026-09-10
Add fcontext support for DEC Alpha (SysV/ELF)

mattst88:
Adds the three fcontext assembly routines for Alpha on Linux (SysV ABI, ELF), plus the b2 and CMake wiring and a doc table entry.

- `src/asm/{make,jump,ontop}_alpha_sysv_elf_gas.S`
- `build/Jamfile.v2`: `asm_sources` alternative for `<architecture>alpha <address-model>64 <abi>sysv <binary-format>elf`
- `CMakeLists.txt`: `alpha` added to `_all_archs`; `CMAKE_SYSTEM_PROCESSOR` is `alpha`, so the existing `IN_LIST` branch matches it the same way riscv64/s390x/loongarch64 are already handled
- `doc/architectures.qbk`: table row

## ABI note

The Alpha calling standard returns a 16-byte struct such as `transfer_t` through a hidden pointer in `$16`, which shifts the declared arguments to `$17`/`$18`. That pointer belongs to the frame that made the call, so it has to travel with the context: each context records the pointer its own suspending call was given, and whoever resumes it stores the `transfer_t` through it. Among the existing ports only ppc32-sysv-linux returns a context through memory this way, so this is the main thing worth reviewing.

`transfer_t` is also passed by value in `$16`/`$17`, which is how a context entered for the first time by `make_fcontext` reads its argument.

## Branch prediction

The transfers end in `ret $31, ($27), 0` rather than `jmp`. The Alpha jump-format opcodes are architecturally identical and differ only in two hint bits driving the return prediction stack. `jump_fcontext` is entered by a `bsr`, which pushes, and the resumption PC is where that call appears to return — so a `jmp` hint orphans one entry per switch, while `ret` pops it. Measured 131.8 -> 128.1 ns/switch (-2.9%) on an EV68AL UP1500, winning all 16 interleaved runs across call depths 0/4/8/16.

(`jsr_coroutine` does not fit despite the name: its push value is architecturally `PC+4`, and fcontext resume points live in the saved context-data, never at `PC+4`.)

## Testing

Assembled, linked and run on real hardware (UP1500, EV68AL, gcc 16.2.0) and under `qemu-alpha`:

- make/jump/ontop round-trip produces the expected switch sequence
- callee-saved registers verified live across `jump_fcontext`: `$9`-`$15` and `$f2`-`$f9`, using values kept live across the call so the compiler allocates them naturally rather than relying on asm clobbers

## Depends on

Architecture detection lands elsewhere; `<architecture>alpha` does not exist without these, so this PR is not usable on its own:

- bfgroup/b2 PR https://github.com/bfgroup/b2/pull/617 — add `alpha` to the `<architecture>` feature
- boostorg/predef PR https://github.com/boostorg/predef/pull/142 — add `alpha` to the architecture check list
- boostorg/config PR https://github.com/boostorg/config/pull/556 — add `checks/architecture/alpha.cpp`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ unset(_default_abi)
 
 ## Arch-and-model
 
-set(_all_archs arm arm64 e2k loongarch64 mips32 mips64 ppc32 ppc64 riscv64 s390x i386 x86_64 combined)
+set(_all_archs alpha arm arm64 e2k loongarch64 mips32 mips64 ppc32 ppc64 riscv64 s390x i386 x86_64 combined)
 
 # Try at start to auto determine arch from CMake.
 if(CMAKE_SYSTEM_PROCESSOR IN_LIST _all_archs)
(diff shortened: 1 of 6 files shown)


olk:
ty

